Issue #180, opened 2026-06-05
tools: withdrawn-message recall not evicted from LatestMessageTracker → all chat-level replies fail with [230011] until next inbound


IS908:
## Summary

When a user withdraws an inbound message, the plugin's `LatestMessageTracker` continues to hold that message_id for up to 10 minutes (or until the next inbound on the same `(chat_id, thread_id)` overwrites it). Within that window, **any reply call that omits `reply_to` fails with `Feishu API [230011]: The message was withdrawn`** — including plain chat-level sends that have no logical relationship to the withdrawn message.

Observed in a P2P chat on 2026-06-05: user sent a second `ping`, then withdrew it. The bot's reply for that turn failed with 230011, and every subsequent retry in the same turn also failed — `reply + reply_to + thread_id`, `reply + thread_id` (plugin auto-fills reply_to), `reply + reply_to` (no thread_id), and **`reply` with only `chat_id` + `text`** (the "fresh chat-level message" path) all returned the same code. Only emitting `[LARK_DEFER]` sentinel let the turn end.

## Root cause chain

**1. Reply tool auto-injects `reply_to` from the tracker** — `src/tools.ts:585-594`

\`\`\`ts
let effectiveReplyTo = reply_to;
if (!effectiveReplyTo && latestMessageTracker) {
  const latest = latestMessageTracker.getLatest(chat_id, thread_id);
  if (latest) {
    effectiveReplyTo = latest.messageId;  // ← injected even when caller passes chat_id only
  }
}
\`\`\`

So a deliberate "send a brand new message to this chat" call (no `reply_to`, no `thread_id`) silently becomes a quoted reply against whatever was most recent.

**2. `LatestMessageTracker` is not invalidated on recall** — `src/channel.ts:262-288` + `:780`

The tracker records every inbound via `handleMessageEvent` with a 10-min TTL. Codebase-wide grep for `recall` / `recalled` / `withdrawn` in `channel.ts` and `index.ts` is empty — there is **no `im.message.recalled_v1` event subscription**, so a withdrawn message stays as the latest entry until TTL expiry or a fresher inbound overwrites it.

Likely also affects:
- `botMessageTracker` (if the bot's *own* message gets withdrawn, the tracked id still appears in reaction-filter checks)
- `recentInboundIds` (`#159` / `#160`) — `revokeAckFor(messageId, markIfMissing=true)` would still mark pending-revoke on a withdrawn id

[...]

PR #184, opened 2026-06-06
fix(channel): doc-comment events 100% dropped in v1.1.0 — wrong payload nesting (v1.1.1, closes #183)

### LINKED ISSUE #183: channel: doc-comment events 100% silently dropped — handleCommentEvent reads from wrong nesting (data.event.notice_meta vs root)
## Summary

`handleCommentEvent` (`src/channel.ts:391`) — added by #182 / closes #181 — reads `data?.event?.notice_meta` and `data?.header?.event_id`, but the Lark Node SDK's `EventDispatcher.register` **auto-unwraps the event envelope** and hands the handler the body directly. So `data.event` is always `undefined`, the early `if (!meta) return;` fires on every event, and **every `drive.notice.comment_add_v1` event is silently dropped**. No log line, no remediation path — the bot looks dead from the user's side.

Confirmed in production on `FsigdRCX3oOQq2xgsptlF9eegLR` (a docx). Three live @-mentions, three silent drops. After adding temp debug to dump `JSON.stringify(data)`, the real shape is:

```json
{
  "schema": "2.0",
  "event_id": "6005621b9176267a45e630a28549de34",
  "token": "",
  "create_time": "1780782786000",
  "event_type": "drive.notice.comment_add_v1",
  "tenant_key": "...",
  "app_id": "cli_a941701876389bd6",
  "comment_id": "7648400110851280613",
  "reply_id": "7648403821734022875",
  "is_mentioned": true,
  "notice_meta": {
    "file_token": "FsigdRCX3oOQq2xgsptlF9eegLR",
    "file_type": "docx",
    "notice_type": "add_reply",
    "from_user_id": { "open_id": "ou_...", "union_id": "...", "user_id": null },
    "to_user_id":   { "open_id": "ou_bot...", "union_id": "...", "user_id": null }
  }
}
```

## Root cause chain

**1. `handleCommentEvent` reads from the wrong nesting** — `src/channel.ts:392, 398`

```ts
const eventId: string | undefined = data?.header?.event_id;  // ← undefined
// ...
const meta = data?.event?.notice_meta;                       // ← undefined
if (!meta) return;                                            // ← every event ends here
```

Compare `handleMessageEvent` at `src/channel.ts:1041`:

```ts
private async handleMessageEvent(data: any): Promise<void> {
  const { message, sender } = data;  // ← direct destructure from root
```

Same SDK, same dispatch path — IM handler reads from root and works; comment handler reads from `.event` / `.header` and silently fails.

**2. Field locations also differ from what the code expects.** Beyond the `.event` nesting bug, several fields the handler later reads from `meta.*` are actually at root of `data`:

| Field          | Code looks at      | Actually at         |
|----------------|--------------------|---------------------|
| `event_id`     | `data.header.*`    | `data.event_id`     |
| `notice_meta`  | `data.event.*`     | `data.notice_meta`  |
| `is_mentioned` | `meta.*`           | `data.is_mentioned` |
| `comment_id`   | `meta.*`           | `data.comment_id`   |
| `reply_id`     | `meta.*`           | `data.reply_id`     |
| `file_token`   | `meta.*`           | `meta.*` ✓          |
| `file_type`    | `meta.*`           | `meta.*` ✓          |
| `from_user_id` | `meta.*`           | `meta.*` ✓          |
| `to_user_id`   | `meta.*`           | `meta.*` ✓          |

So even fixing the `.event` access alone isn't enough — `meta.is_mentioned !== true` (line 402), `meta.comment_id` (line 411), `meta.reply_id` (line 412) all need to move to `data.*`.

**3. Smoke test mocks the wrong shape, so it can't catch this.** `scripts/comment-event-smoke.ts:38-52`:

```ts
return {
  header: { event_id, event_type: 'drive.notice.comment_add_v1' },
  event: {
    notice_meta: {
      file_type: 'docx',
      file_token,
      comment_id,        // ← inside notice_meta
      reply_id,          // ← inside notice_meta
      is_mentioned,      // ← inside notice_meta
      from_user_id: { open_id: from_open_id },
      to_user_id:   { open_id: to_open_id },
    },
  },
};
```

That matches what `handleCommentEvent` reads but not what the SDK actually delivers, so the smoke test is green while production is 100% broken. Classic mock-vs-reality divergence — same family as #180's tracker assumption.

## Reproduction

1. Feishu app with the `drive.notice.comment_add_v1` event subscription enabled (per `CLAUDE.md` Doc-comment setup).
2. Open a doc the bot is a collaborator on, post a comment that @-mentions the bot.
3. Tail `~/.claude/channels/lark/debug.log`. You'll see:

```
[ts] [channel] Event received: drive.notice.comment_add_v1
```

…and **nothing else**. No \"Doc comment from … rejected by whitelist\", no envelope build, no enqueue. Whitelist rejection logs (`src/channel.ts:427-431`) confirm the dropout is upstream of the whitelist check.

## Fix shape (uncontroversial — single PR)

In `src/channel.ts`:

```ts
const eventId: string | undefined = data?.event_id;
// ...
const meta = data?.notice_meta;
if (!meta) return;

if (data.is_mentioned !== true) return;
if (meta.to_user_id?.open_id !== deps.botOpenId) return;
if (meta.from_user_id?.open_id === deps.botOpenId) return;

const fileToken: string = meta.file_token;
const commentId: string = data.comment_id;
const replyId: string | undefined = data.reply_id;
const fileType: string = meta.file_type;
const fromOpenId: string = meta.from_user_id.open_id;
```

And in `scripts/comment-event-smoke.ts:makeEvent`, replace the wrapped fake with a flat-shape one matching the real SDK payload above (the 4 misplaced fields). Without that, future regressions in the same direction will still pass smoke.

## Why this slipped through review

PR #182 review focused on the security and identity-binding semantics (`passesDocCommentWhitelist`, per-comment `setCaller`, the `__terminal__` rejection in `reply_doc_comment`). The wire-format assumption never had an integration touchpoint — there's no \"send a real Feishu event through `EventDispatcher.register`\" smoke. The unit-level smoke just calls `handleCommentEvent({header, event: {...}}, deps)` directly with a hand-built mock, so the mismatch was invisible.

Two adjacent gaps worth thinking about as part of the fix:

- The IM handler (`handleMessageEvent`) was implemented years ago by reading the SDK's actual payload at runtime — that's why it has the right shape. There's no shared type or schema between IM and doc-comment handlers; both should probably reference whatever Lark SDK types exist for these events (or fail loudly if a critical field is missing on receipt, e.g. log `Object.keys(data)` once on first event of each type).
- Silent `return` in event handlers is a footgun. Every drop point in `handleCommentEvent` (`!meta`, `is_mentioned !== true`, `to !== bot`, `from === bot`) was silent in v1.1.0 — only the whitelist check logs. A one-line `debugLog` in each branch (`reason=...`) costs nothing and would have flagged this within minutes of the first @-mention.

## Related

- #181 — original feature spec.
- #182 — implementation PR that introduced this bug.
- #180 — same family (mock shape diverges from production reality, silent failure mode).

IS908:
## Summary

Hotfix for #183 — doc-comment @bot events (added by #182 in v1.1.0) were silently dropped on 100% of inbound events in production. The Lark Node SDK's `EventDispatcher.register` auto-unwraps the envelope and delivers the event body directly, but `handleCommentEvent` read from `data.header.event_id` / `data.event.notice_meta` — so `data.event` was always `undefined`, the early `if (!meta) return` fired on every event, and no doc comment ever reached Claude. The smoke factory `makeEvent` duplicated the buggy nesting, so all 20 smoke cases were green while production was 100% broken — same family as #180.

## Field mapping (issue #183)

| Field | Was reading | Should read |
|---|---|---|
| `event_id` | `data.header.event_id` | `data.event_id` |
| `notice_meta` | `data.event.notice_meta` | `data.notice_meta` |
| `is_mentioned` | `meta.is_mentioned` | `data.is_mentioned` |
| `comment_id` | `meta.comment_id` | `data.comment_id` |
| `reply_id` | `meta.reply_id` | `data.reply_id` |
| `file_token` | `meta.file_token` | `meta.file_token` (unchanged) |
| `file_type` | `meta.file_type` | `meta.file_type` (unchanged) |
| `from_user_id` | `meta.from_user_id` | `meta.from_user_id` (unchanged) |
| `to_user_id` | `meta.to_user_id` | `meta.to_user_id` (unchanged) |

## Commits

- `d2abcb8` — `fix(channel): read doc-comment event fields from real SDK payload shape (#183)`. Rewrote field extraction in `handleCommentEvent` and rewrote the `makeEvent` factory in `scripts/comment-event-smoke.ts` to mirror the real SDK shape. Also adds `debugLog` to every silent-return path (missing `notice_meta`, `is_mentioned=false`, `to_user_id` mismatch, self-loop) — adjacent gap also flagged in #183.
- `3facb84` — `chore(release): bump version to 1.1.1 for #183 fix`. Bumps 6 version sites + CHANGELOG.

## Verification

- `npm run typecheck` — clean
- `bash scripts/test.sh` — all green, including `comment-event-smoke 20/20` (the rewritten mock did not break any semantic assumption — every case still exercises the intended branch)
- `npm start -- --dry-run` — exits 0, stdout empty
- Mental trace confirms the new `debugLog` lines fire as expected: smoke output now prints "is_mentioned=false / to_user_id != bot / from bot itself" diagnostics for cases 2/3/4.

## Acceptance

M1 from the original spec (owner @-mentions bot in a docx comment, observes a Claude turn) should now actually fire. Operator should run this against a live tenant before declaring victory — the smoke suite still calls `handleCommentEvent` directly with hand-built mocks; integration smoke through `Lark.EventDispatcher.register` is logged as a follow-up in the CHANGELOG.


IS908:
## First-round review on PR #184 hotfix

**0 Critical, 1 Important, 4 Minor.** The code fix is correct field-for-field against the #183 payload; all version bumps complete; no other handlers have the same nesting bug. Verdict: approve with the docs sweep.

### Critical / Important

**I-1. Docs still advertise "v1.1.0+" for the doc-comment feature — but v1.1.0 was 100% broken**

[CLAUDE.md:91](https://github.com/IS908/claude-lark-plugin/blob/fix/doc-comment-payload-shape/CLAUDE.md#L91), README.md (3 places at ~248, 364, 365), README_CN.md (3 places at ~83, 229, 355, 356) all say the doc-comment feature is available "v1.1.0+". A user reading the README sees v1.1.0 advertised, installs it, gets nothing — the whole point of v1.1.1 is that this is the version where it actually works. Future support load: "Why didn't it work on v1.1.0?" / "Right, that was the broken one."

Should bump every doc-comment-only "v1.1.0+" reference to "v1.1.1+". Mechanical sed, ~7 line-edits.

### Minor (non-blocking)

- **M-1.** Mock includes `notice_type: reply_id ? 'add_reply' : 'add_comment'` but the handler doesn't read it (uses presence/absence of `reply_id` to branch). Dead weight today but accurate to wire shape.
- **M-2.** Default mock passes `reply_id: undefined` for `add_comment` cases; production payload sends `reply_id: null`. Both are correctly handled by the `if (replyId)` truthiness check, so no bug. For full shape fidelity, `null` would mirror wire format better.
- **M-3.** New debugLog at line 420 echoes `to_user_id.open_id`; sourced from signed Feishu event, consistent with IM whitelist log's open_id echo. No action needed.
- **M-4.** CHANGELOG entry position / format / accuracy all good.

### Confirmed correct after deep probe

- **Field mapping verified field-for-field** against the production dump in #183 (all 9 fields).
- **Version bump complete** at all 5 sites (`package.json`, `package-lock.json` top + `packages[""]`, `.claude-plugin/plugin.json`, `.claude-plugin/marketplace.json`, `src/index.ts`). No stray `1.1.0` strings remain in live code.
- **No other handler has the same nesting bug.** `handleMessageEvent` destructures `{message, sender} = data` from root; `handleReactionEvent` reads `data.message_id` / `data.reaction_type` / `data.operator_type` / `data.user_id.open_id` from root. Neither uses `data.event` or `data.header`. Bug was isolated.
- **All silent returns now logged** except dedup early return — correct, since dedup hits are high-volume.
- **debugLog format consistent** with existing `[channel] ...` prefix + event_id-parameterized convention.
- **Smoke 20/20** passes; walked each case mentally — semantic intent preserved after the factory rewrite.
- **Production extras** (`schema`, `token`, `create_time`, `tenant_key`, `app_id`) included in new mock — full shape fidelity. Catches future regressions of this class.
- **typecheck clean**; full `bash scripts/test.sh` green.

### Summary

|---|---|---|
| Critical | 0 | — |
| Important | 1 | I-1: docs reference v1.1.0+ for a feature that's only really available in v1.1.1+ |
| Minor | 4 | mock fidelity nits, format consistency |

The code is correct and tightly scoped. The docs sweep (I-1) is the only thing worth fixing in the same PR — 7 line-edits across 3 files. After that lands, ready to ship.

IS908:
Round-1 I-1 addressed in commit `ecc2721`. 7 doc-references updated from v1.1.0+ → v1.1.1+ across CLAUDE.md / README.md / README_CN.md. Historical CHANGELOG references left intact. Tests green.

IS908:
## Second-round review on PR #184

**CONVERGED with 1 minor inconsistency.**

### Important / Critical
None.

### Minor
**M-1.** Two sibling references about LRU sizing inherited the v1.1.0 framing the round-1 sweep didn't reach:

- [README.md:284](https://github.com/IS908/claude-lark-plugin/blob/fix/doc-comment-payload-shape/README.md#L284) — "Doc-comment events (post-v1.1.0) key per-comment..."
- [README_CN.md:265](https://github.com/IS908/claude-lark-plugin/blob/fix/doc-comment-payload-shape/README_CN.md#L265) — "文档评论事件（v1.1.0 起）按 comment_id 分桶..."

A user reading these next to the freshly-bumped v1.1.1+ headline markers sees contradictory signals — "feature available since 1.1.0" vs "feature working since 1.1.1". Semantically the LRU-sizing advice is only operationally true from v1.1.1 (in v1.1.0 every event was dropped, so `setCaller("doc:<token>", comment_id, ...)` never ran, LRU saw zero doc-comment entries, the rationale doesn't apply).

Not blocking — the implementer flagged this scope decision in their own report. Two-line bump for full consistency.

### Confirmed correct
- All 7 round-1 bumps applied cleanly, no off-target edits.
- CHANGELOG 1.1.1 entry explicitly says "Doc-comment events 100% silently dropped in v1.1.0" — sweep premise validated.
- Reverse-check: no bump overshot. All 7 references correctly identify v1.1.1 as the first working version.
- Source-tree grep clean for misleading v1.1.0 markers. `src/channel.ts:396` correctly says "Pre-v1.1.1 code read from..."
- typecheck + full test.sh green.

### Verdict
Will fix M-1 in a follow-up commit on this PR for completeness, then this is ready.

IS908:
Round-2 M-1 addressed. Both LRU-sizing notes bumped to v1.1.1+ for full consistency. Loop should converge.

IS908:
## Third-round review — **CONVERGED**, ready to merge

- **Commit `9b3acc0`** both edits clean, no off-target changes.
- **Final consistency sweep**: grep `*.md` and source files for any remaining misleading v1.1.0 feature-availability markers → **zero hits**. Only remaining v1.1.0 references are in CHANGELOG (correctly describing v1.1.0 as the broken version).
- **Tests**: typecheck clean, full `scripts/test.sh` green (113 hook + 20 comment-event + 15 reply-doc-comment + 10 comment-elements + all pre-existing).
- **PR state**: `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.
- **PR title** matches repo convention `(v1.1.1, closes #183)` — squash will append `(#184)`.

### Cumulative loop stats (PR #184)

| Round | Found | Class |
|---|---|---|
| #183 (external) | Wire-format mismatch — 100% silent drop in production | Critical production bug |
| 1 (my review) | I-1: docs say "v1.1.0+" for the broken feature | Important docs accuracy |
| 2 (my review) | M-1: 2 LRU-sizing references still "post-v1.1.0" / "v1.1.0 起" | Minor consistency |
| 3 (my review) | **CONVERGED** | — |

**Ready for merge + v1.1.1 release.**

PR #186, opened 2026-06-07
fix(channel): anchored doc-comments 404 in pre-fetch — switch to list endpoints (v1.1.2, closes #185)

### LINKED ISSUE #185: channel: handleCommentEvent body pre-fetch 404s for anchored comments (is_whole=false) — fileComment.get unsupported, use fileComment.list/fileCommentReply.list
## Symptom

For doc-comment events on *anchored* (划词/quoted) comments — `is_whole=false` with a `quote` field — `handleCommentEvent` in `src/channel.ts:464` always returns `fetchError`, so the envelope reaches Claude with an empty `<body>` and a visible `<fetch_error>Request failed with status code 404</fetch_error>`. Claude can't see what the user wrote and has to ask them to repeat.

Whole-doc comments (`is_whole=true`, no `quote`) are unaffected — the GET endpoint works for those.

## Repro (verified live against open.feishu.cn 2026-06-07)

A real anchored comment thread on a docx:
- `file_token=FT7udZ6dxohd81xpeFOcgAhGnct`
- `comment_id=7648278408253672683` (`is_whole=false`, `quote="Boeing"`)
- `reply_id=7648486568478395616` (new "ping" reply that triggered the event)

| Endpoint | Result |
|---|---|
| `GET /drive/v1/metas/batch_query` (doc meta) | 200 ✔ — bot has read access |
| `GET /drive/v1/files/:token/comments?file_type=docx` (LIST comments) | 200 ✔ — returns the thread including the new reply |
| `GET /drive/v1/files/:token/comments/:comment_id/replies?file_type=docx` (LIST replies) | 200 ✔ — returns all 4 replies |
| `GET /drive/v1/files/:token/comments/:comment_id?file_type=docx` (**our call**) | **404** `code=1069307, msg="not exist"` |

Same `comment_id`, same `file_type`, same token — only the single-fetch endpoint returns "not exist". This is an upstream Feishu API limitation on anchored comments, not a permission/scope issue.

## Root cause

`src/channel.ts:464-485` uses `deps.client.drive.fileComment.get(...)` to pre-fetch both the parent body and (when `replyId` is set) the target reply body. That endpoint maps to `GET /drive/v1/files/:file_token/comments/:comment_id` (confirmed at `node_modules/@larksuiteoapi/node-sdk/es/index.js:39398`) which doesn't support `is_whole=false` comments.

Anchored comments are the *typical* UX for "@-bot in a docx" — users highlight text and @-mention. So this currently silently breaks the common case while only the rare whole-doc path works.

## Fix sketch

Replace the single `fileComment.get` call with two list calls that both work:

1. `fileComment.list({ path:{file_token}, params:{file_type, page_size:N} })` → find the matching `comment_id` to get `quote` and `reply_list.replies[0].content` (parent body, when `replyId` is set).
2. `fileCommentReply.list({ path:{file_token, comment_id}, params:{file_type, page_size:N} })` → all replies; pick the one whose `reply_id === replyId` for the target body. (For `add_comment` with no `replyId`, replies[0] is the comment body itself.)

Open question for the fix: paginating `fileComment.list` to find a specific `comment_id` is O(N) in the doc's comment count. Two reasonable options:
- (a) Always call `fileCommentReply.list` (cheap, scoped to one thread) for body; only call `fileComment.list` when we need `quote` and `parentBody`. Most events are `add_reply` so this is the hot path.
- (b) Keep `fileComment.get` as the first attempt; fall back to list on 1069307. Costs an extra round-trip on the common case but minimizes diff.

I'd default to (a) — separates "body of the event" (always available via replies endpoint) from "thread context" (only needed for `parentBody`/`quote`).

## References

- `src/channel.ts:457-485` — current pre-fetch code
- `node_modules/@larksuiteoapi/node-sdk/es/index.js:39398` — SDK endpoint for `fileComment.get`
- Feishu API docs: https://open.feishu.cn/document/server-docs/docs/CommentAPI/get (single get — fails on anchored), `…/CommentAPI/list` (works), `…/file-comment-reply/list` (works)
- Originally diagnosed when the @bot mention thread on `FT7udZ6dxohd81xpeFOcgAhGnct` arrived with empty body

IS908:
## Summary

Closes #185. In v1.1.1, `handleCommentEvent` used `drive.fileComment.get` for the pre-fetch, but that endpoint returns 404 (`code=1069307`, `msg="not exist"`) for any comment where `is_whole=false` — the typical UX for @-mentioning the bot inside a docx (user highlights text + types `@bot`). Result: Claude saw empty `<body>` and `<fetch_error>Request failed with status code 404</fetch_error>` for the common case while only the rare whole-doc path worked. Replaced with parallel `drive.fileCommentReply.list` (for body / parent body) + `drive.fileComment.list` (for quote). Both list endpoints work for anchored AND whole-doc comments.

## Endpoint matrix (from live #185 testing)

| Endpoint | `is_whole=true` (whole-doc) | `is_whole=false` (anchored) |
|---|---|---|
| `GET .../comments/:comment_id` (`fileComment.get` — pre-fix) | 200 ✔ | **404** `code=1069307` ✘ |
| `GET .../files/:token/comments` (`fileComment.list` — post-fix, for `quote`) | 200 ✔ | 200 ✔ |
| `GET .../files/:token/comments/:cid/replies` (`fileCommentReply.list` — post-fix, for body) | 200 ✔ | 200 ✔ |

## Implementation notes

- **`src/channel.ts:handleCommentEvent`**: single GET → parallel `Promise.all([fileCommentReply.list, fileComment.list])`. Per Feishu's data model (verified live in #185), `items[0]` from `fileCommentReply.list` IS the original comment body, so existing body/parentBody extraction logic is preserved in spirit.
- **`CommentEventDeps.client` interface**: removed `fileComment.get`, added `fileComment.list` and `fileCommentReply.list`. SDK type defs verified against `node_modules/@larksuiteoapi/node-sdk/types/index.d.ts:212204` and `:212338`.
- **`scripts/comment-event-smoke.ts`**: rewrote `makeDeps` default factory and all 5 ad-hoc per-case clients (cases 6, 7, 8, 12, 14) to mock the new endpoints. Added case 19 explicitly exercising the anchored-comment path (`is_whole=false`) — verifies both list calls fire, `quote` surfaces in `<selected_text>`, body in `<body>`. All 20 prior cases preserve their semantic intent.

## Test plan

- [x] `npm run typecheck` clean
- [x] `npx tsx scripts/comment-event-smoke.ts` — PASS: 21 cases (was 20; +1 for #185)
- [x] `bash scripts/test.sh` — all tests pass
- [x] `LARK_APP_ID=cli_test LARK_APP_SECRET=secret npm start --silent -- --dry-run` — exit 0, stdout empty
- [ ] Manual live verification against a real anchored docx comment (post-merge)

## Known follow-up (intentionally out of scope)

Pagination capped at `page_size=100` for both list calls. If a doc has >100 comments or a thread has >100 replies, the matching item may not be on page 1 — body falls back to `<body unknown="true">` (existing add_reply path) or `quote` is silently omitted. Acceptable for v1.1.2 hot-path fix; `page_token` loop is a follow-up tracked in CHANGELOG.


IS908:
## First-round review on PR #186 (#185 hotfix)

**0 Critical, 2 Important, 5 Minor.** SDK shapes verified line-by-line against type defs; case 19 is a real regression guard. Two substantive findings worth raising before merge.

### Important

**I-1. `Promise.all` couples body delivery to quote-list success.** [src/channel.ts:489-498](https://github.com/IS908/claude-lark-plugin/blob/fix/comment-anchored-prefetch/src/channel.ts#L489-L498)

The body is the actionable payload (what the user wrote). The quote is auxiliary context (highlighted text). Coupling their futures via `Promise.all` means a transient 429 / scope mismatch / API blip on `fileComment.list` (quote source) wipes a perfectly fine `fileCommentReply.list` result (body source) — every event in that window surfaces `<fetch_error>` instead of a working body+missing-quote envelope.

Cheap fix: switch to `Promise.allSettled`, render `<body>` from the resolved replies result, render `<selected_text>` only when the comments result resolved AND the matching comment had a quote. Leave `<fetch_error>` for the case where BOTH list endpoints fail.

```ts
const [repliesResult, commentsResult] = await Promise.allSettled([
  deps.client.drive.fileCommentReply.list({ ... }),
  deps.client.drive.fileComment.list({ ... }),
]);
const replies: any[] = repliesResult.status === 'fulfilled' ? (repliesResult.value?.data?.items ?? []) : [];
const comments: any[] = commentsResult.status === 'fulfilled' ? (commentsResult.value?.data?.items ?? []) : [];

if (repliesResult.status === 'rejected' && commentsResult.status === 'rejected') {
  fetchError = (repliesResult.reason?.message ?? String(repliesResult.reason));
} else if (repliesResult.status === 'rejected') {
  fetchError = `replies list failed: ${repliesResult.reason?.message ?? String(repliesResult.reason)}`;
}
// quote-only failure → silently omit <selected_text>, no fetch_error
```

**I-2. Live API verification of `fileComment.list` for anchored comments.** Issue #185 already demonstrates this experimentally (`LIST` endpoints return 200 for the same anchored comment that `GET` 404'd on). The PR body / CHANGELOG should explicitly cite this repro to short-circuit the "another wire-format mismatch shipped" failure mode. Suggested addition to CHANGELOG implementation note: "Endpoint behavior live-verified against open.feishu.cn 2026-06-07 per #185 — `fileComment.list` returns `is_whole=false` items by default, no explicit param needed."

### Minor

- **M-1.** [CLAUDE.md:44](https://github.com/IS908/claude-lark-plugin/blob/fix/comment-anchored-prefetch/CLAUDE.md#L44) still describes pre-fetch as `drive.fileComment.get`. Should mention the two new list endpoints. Docs-lag-code class — 5th file surface this cycle.
- **M-2.** `v1.1.1+` markers in README/README_CN (8 locations: README.md:248/284/364/365 + README_CN.md:83/229/265/355/356) should be `v1.1.2+` for the anchored-comment feature. Users reading the README at HEAD treat v1.1.1 as a safe pin; it isn't (anchored case is the typical UX and was broken there).
- **M-3.** Case 7 (`failingClient`) makes BOTH list endpoints throw — `Promise.all` rejects on first failure so the test would pass with only one throwing. After I-1 fix (allSettled), this case won't even catch a partial-failure regression. Add a case 7c (replies succeeds + comments fails → body renders, quote omitted, no `<fetch_error>`) to lock in chosen behavior.
- **M-4.** Case 6 (`replyClient`) mocks `quote: 'q'` but doesn't assert `<selected_text>` rendering. Cheap tightening — case 19 already covers full quote path so not urgent.
- **M-5.** `page_size=100` hardcoded. No env knob (`LARK_DOC_COMMENT_PREFETCH_PAGE_SIZE`?). Documented as known follow-up, acceptable for hot-path fix.

### Confirmed correct

- SDK shapes match type defs at `:212204` (fileComment.list) and `:212338` (fileCommentReply.list).
- Response destructuring: `data.items` (not the old `data.reply_list.replies`).
- `extractText` handles both `{text}` and `{elements:[...]}` shapes correctly.
- Body/parent/quote logic: add_comment, add_reply happy, add_reply missing all behave per spec; `<body unknown="true">` fallback preserved.
- `CommentEventDeps` interface cleanly removed `fileComment.get`. No leftover call sites of the old endpoint in production code (`src/tools.ts` `DocCommentClient` is `create` endpoints only, unrelated).
- Version bumped at all 5 sites.
- CHANGELOG entry honest about page_size follow-up.
- Case 19 verifies BOTH endpoints fire AND `<body>` AND `<selected_text>` render — appropriate regression guard.

### Summary

|---|---|---|
| Critical | 0 | — |
| Important | 2 | I-1: Promise.all atomicity wipes body on quote-list failure; I-2: cite live-API verification from #185 |
| Minor | 5 | docs-lag-code, smoke coverage gaps, env knob |

Recommend addressing I-1 + I-2 + M-1 + M-2 in a single follow-up commit before merge.

IS908:
Round-1 findings addressed in commit `1c55cc11c8150069a640ac92d0245aa67cb4e432`:

- **I-1** (Promise.allSettled): partial failures no longer wipe body; new case 7c locks in behavior.
- **I-2**: CHANGELOG implementation note now cites live-API verification from the #185 repro.
- **M-1**: CLAUDE.md doc-comment flow description updated to reflect current endpoints.
- **M-2**: 9 v1.1.1+ markers across README.md (4) + README_CN.md (5) bumped to v1.1.2+ — v1.1.1 worked only for whole-doc; anchored is the typical UX.
- **M-3**: case 7c added — quote-only failure renders body without fetch_error.
- **M-4**: case 6 now asserts <selected_text>q</selected_text>.
- **M-5** (page_size env knob): deferred as documented follow-up.

`bash scripts/test.sh` green; smoke now 22/22.

IS908:
## Second-round review on PR #186

**0 Critical, 0 Important, 4 Minor.** `Promise.allSettled` mechanics are correct across all 4 branches; case 7c assertions tight; case 6 quote rendering closes coverage cleanly; CHANGELOG cites #185 live-API verification; README/README_CN bumps complete. Verdict: **CONVERGED with minor nits**.

### Critical / Important
None.

### Minor

**N-1.** [CLAUDE.md:91](https://github.com/IS908/claude-lark-plugin/blob/fix/comment-anchored-prefetch/CLAUDE.md#L91) still says "Doc-comment events (v1.1.1+)" while all 9 README/README_CN markers were bumped to v1.1.2+ in round-1. Inconsistency — a user reading both docs gets contradictory signals about which version the feature actually works on. Bump for consistency.

**N-2.** [scripts/comment-event-smoke.ts:187-190](https://github.com/IS908/claude-lark-plugin/blob/fix/comment-anchored-prefetch/scripts/comment-event-smoke.ts#L187-L190) case 7's inline comment still says "via Promise.all; either throwing rejects the await and triggers the catch" — no longer matches the code after allSettled refactor. Behavior is preserved, but future readers get misled. 1-line edit.

**N-3.** Partial-failure paths in [src/channel.ts:515-522](https://github.com/IS908/claude-lark-plugin/blob/fix/comment-anchored-prefetch/src/channel.ts#L515-L522) — quote-only failure (`commentsResult.status === 'rejected'`) is silently absorbed with no `debugLog`, so operators have no signal that `fileComment.list` is degraded. Existing handler logs every other drop path. 1-line `debugLog` would close the observability gap.

**N-4.** Case 7c is placed at the bottom of the smoke file rather than next to cases 7 / 7a / 7b (all "pre-fetch failure" family). Maintainability nit — purely cosmetic. Skip.

### Confirmed correct
- `Promise.allSettled` correctness: walked all 4 branches; each produces the documented envelope. Old `try/catch` removed cleanly.
- Case 7c asserts NO `<fetch_error>` + `<body>` + NO `<selected_text>` — all three tight.
- Case 6 line 295 asserts `<selected_text>q</selected_text>` exactly.
- README/README_CN `v1.1.1+` grep returns nothing (all 9 bumped).
- CHANGELOG cites #185 repro for `fileComment.list` returning `is_whole=false` items + `fileCommentReply.list` original-comment-as-items[0] behavior.
- Cases 7 / 7a / 7b behavior preserved (both-reject still triggers fetchError; escape-ordering unaffected).
- Edge case (data.items undefined vs []): both yield safe `<body unknown="true">` envelope.
- Recorder array naming consistent across declarations / pushes / assertions.
- Tests all green; PR `MERGEABLE` + `CLEAN`.
- `check-doc-surfaces.sh` only known pre-existing false-positives.

### Verdict
Will fix N-1 / N-2 / N-3 in one small follow-up commit, then this should converge cleanly.

IS908:
Round-2 nits N-1 + N-2 + N-3 addressed. N-4 skipped as cosmetic.

- N-1: CLAUDE.md:91 bumped v1.1.1+ → v1.1.2+ for consistency with READMEs.
- N-2: case 7 inline comment rewritten — no longer references the removed Promise.all mechanism.
- N-3: 3 debugLog branches added to handleCommentEvent for partial-failure visibility (replies-only reject, comments-only reject, both reject). Operators now have a signal when fileComment.list is degraded.

`bash scripts/test.sh` green; smoke still 22/22.

IS908:
## Third-round review — **CONVERGED, ready to merge**

- **N-1** verified: CLAUDE.md:91 marker bumped to v1.1.2+, consistent with all 9 README/README_CN markers.
- **N-2** verified: case 7 comment now names `Promise.allSettled` + cross-references case 7c. Historical `Promise.all` mentions elsewhere (smoke:191/553) are deliberate context.
- **N-3** verified: all 3 debugLog branches present with event_id + comment_id; comments-only-reject branch correctly omits fetchError (auxiliary).
- **Sweep**: `grep -nE "v1\.1\.1\+" CLAUDE.md README.md README_CN.md` → zero hits.
- **Tests**: typecheck clean, `scripts/test.sh` all suites green (113/113 hook + 22/22 comment-event + 15/15 reply-doc-comment + 10/10 comment-elements + 14/14 distill).
- **PR state**: `MERGEABLE` + `CLEAN`, title matches `(v1.1.2, closes #185)` convention.

### Cumulative loop stats (PR #186)

| Round | Found | Class |
|---|---|---|
| #185 (external) | Anchored is_whole=false comments 404 on fileComment.get | Production bug — typical UX broken |
| 1 (my review) | I-1 Promise.all atomicity wipes body; I-2 cite live verification; 5 Minor | Important x2 + docs/coverage |
| 2 (my review) | N-1 CLAUDE.md inconsistent v1.1.1+; N-2 stale Promise.all comment; N-3 silent partial failure | Minor docs/observability |
| 3 (my review) | **CONVERGED** | — |

**Ready for merge + v1.1.2 release.**

Issue #190, opened 2026-06-12
memory: idle-triggered /compact + fallback /clear with Agent/Todo state persistence to bound accumulated history cost


IS908:
## Background

The plugin's Claude Code main loop is a single long-running interactive session shared by every chat / thread / cronjob. Conversation history accumulates indefinitely until Claude Code's built-in auto-compaction kicks in near context limits. By that point the megasession may already carry tens of thousands of tokens of stale memory_context, replies to long-resolved questions, and Agent / TodoWrite state from finished tasks.

After a > 5 min idle gap the next inbound pays cache-write (~125% of input price) on the entire accumulated payload. Even with #189 (per-turn injection trimming) shrinking the per-turn slab, the **already-committed** history is fixed once written.

## Problem

We want a way to actively shrink already-accumulated history at safe boundaries (idle moments when nothing is in flight) rather than wait for context-limit auto-compaction. Concretely:

- 30 min idle → trigger \`/compact\` (Claude Code's built-in summarization) so the next burst pays cache write on a compressed payload instead of full history
- 6 h idle OR token-usage > threshold → trigger \`/clear\` (full reset) and rely on memory_context to re-establish grounding next turn
- Both branches must **preserve in-flight state**: Agent processes, TodoWrite items, pending plan steps, partially-prepared replies — otherwise we trade tokens for silently orphaned work

## Proposed flow

State machine, all configurable, defaults preserve current behavior:

\`\`\`
on tick (every N seconds):
  if any chat has pending Stop-hook obligation        → skip   (work in flight)
  if any cronjob currently executing                  → skip
  if any Agent process spawned and not yet completed  → skip
  if idle < LARK_IDLE_COMPACT_MS (default 30m)        → skip
  if idle >= LARK_IDLE_CLEAR_MS (default 6h)
     OR session_tokens > LARK_TOKEN_CLEAR_THRESHOLD   → CLEAR branch
  else                                                → COMPACT branch

COMPACT branch:
  snapshot in-flight state (Todos, pending Agents, plan) to ~/.claude/channels/lark/session-state.json
  inject /compact into main loop
  log to audit
  next inbound: enrichWithMemory adds a session_restore block summarising the snapshot

CLEAR branch:
  snapshot all of the above
  inject /clear into main loop
  log to audit
  next inbound: session_restore block is injected as a HOT memory_context section
  state file is removed after Claude acknowledges restoration
\`\`\`

### State snapshot contents (best-effort)

[...]

PR #191, opened 2026-06-12
feat(memory): content-hash dedup of memory_context injection on hot threads (v1.3.0, closes #189)

### LINKED ISSUE #189: memory: reduce per-turn memory_context injection on thread follow-ups to cut accumulated history token cost
## Background

`LarkChannel.enrichWithMemory` (src/channel.ts:1475) currently injects the same shape of `<memory_context>` envelope on **every inbound message**:

1. `profile` (self) — always loaded, full content (both public + private tiers since caller == owner)
2. `mentioned_profile` (each @-mention) — always loaded, public tier only
3. `thread_episode` × N — semantic search filtered by `LARK_MIN_SEARCH_SCORE` (default 0.3), capped at `LARK_MAX_SEARCH_RESULTS` (default 2), each capped at `LARK_EPISODE_INJECT_CAP_BYTES` (default 2 KB)
4. `chat_episode` × N — same caps
5. `skill` × N — same caps

The injection runs on the Claude Code main loop's interactive session (single long-running REPL). The main loop already keeps prior turns in its conversation history, so on a thread follow-up the **previous turn's memory_context is still in context** — re-injecting nearly-identical blocks on the next turn is mostly duplication that lands in cache writes whenever the 5-minute prompt-cache TTL expires.

## Problem

Multi-turn cost on the plugin's shared megasession is dominated by accumulated history (every prior turn's memory_context + user text + assistant reply stays in the conversation array). After a > 5 min idle gap, the next turn pays cache-write (~125% of input price) on the entire accumulated payload. Memory_context contributes a fixed-ish slab to every turn that we can shrink without losing semantic value, because **the LLM already saw it on the prior turn**.

Concrete back-of-envelope on a typical follow-up turn in this repo's P2P bot:
- profile (self, both tiers): ~1.5–3 KB
- 2 thread_episodes (capped at 2 KB each): up to 4 KB
- 2 chat_episodes (capped at 2 KB each): up to 4 KB
- 2 skills with description: ~0.5 KB
- envelope overhead + preamble: ~0.5 KB

→ ~7–10 KB per turn just for memory_context, repeated each turn, all of it sitting in conversation history forever (until Claude Code auto-compaction kicks in).

Over 20 thread turns this is ~150–200 KB of mostly-redundant injection that the model already saw early in the thread.

## Current behavior

- `enrichWithMemory` does the same full injection on every turn regardless of:
  - Whether the prior turn (same thread, same user) just received the same blocks
  - Whether `msg.threadId` is set (we always also inject chat-wide episodes, even when thread context is more relevant)
  - Whether the message looks like an operational query (skills useful) vs a Q&A follow-up (skills mostly noise)
- Knobs that exist today: `LARK_MIN_SEARCH_SCORE`, `LARK_MAX_SEARCH_RESULTS`, `LARK_EPISODE_INJECT_CAP_BYTES` — all global, all time-invariant.

## Proposed direction

Add **context-aware injection** that downscales memory_context when the same conversation thread is hot. Strawman:

### 1. Follow-up suppression on hot threads

If the current inbound is a follow-up in an active thread (same `chat_id` + `thread_id`, prior inbound less than \`LARK_MEMORY_FOLLOWUP_WINDOW_MS\` ago — proposed default 5 min, aligned with prompt-cache TTL), skip or shrink redundant blocks:

| Block | First turn in thread | Follow-up (within window) |
|---|---|---|
| profile (self) | full | omit (already in conversation history) |
| mentioned_profile | full | omit |
| thread_episode | top-N | omit (the running conversation IS the thread context) |
| chat_episode | top-N | top-1 only, and only if score > stricter cutoff |
| skill | top-N | omit unless current message has obvious operational intent (keyword filter) |

The first turn in a thread bears full memory_context to establish grounding; subsequent turns rely on the conversation-history copy that's already in cache.

### 2. Thread-scope vs chat-scope episode trimming

When `msg.threadId` is set, the user is clearly in a focused sub-conversation. Reduce `chat_episode` injection to top-1 (or drop entirely) when at least one `thread_episode` survives the score filter — the wider chat history is much less likely to be relevant than the thread's own past.

### 3. Skill injection becomes opt-in by query signal

Skills are only useful when the user is asking the bot to *do* something. For pure Q&A / discussion turns they are noise that costs tokens every turn. Heuristic: inject skills only when the inbound contains operational triggers (imperative verbs, slash patterns, tool/CLI mentions, attachment refs) OR explicit \`/skill\` invocation — otherwise skip.

### 4. New knobs (all backward-compatible)

- \`LARK_MEMORY_FOLLOWUP_WINDOW_MS\` — default 300_000 (5 min); 0 disables follow-up suppression
- \`LARK_MEMORY_THREAD_PRIORITIZE\` — default true; when set, thread_episode hits cause chat_episode count to drop
- \`LARK_MEMORY_SKILLS_OPT_IN\` — default false (preserve current behavior); when true, skills are only injected on operational-signal turns

Roll out behind flags so the cost/benefit can be measured per-deployment before flipping defaults.

## Open questions

1. **Where to track \"prior inbound timestamp per thread\"?** Easiest: in-memory map keyed by \`\${chatId}:\${threadId}\`, populated by \`handleMessageEvent\`. Loses state across restarts — acceptable since this is a cost-optimization, not correctness.

2. **What's the right \"operational signal\" heuristic for #3?** Initial candidates: matches \`/skill_name\`, contains imperative verbs in \`en|zh\` whitelist, contains a CLI/tool token. Want this to be precision-leaning (false negative → carry skills like today; false positive → save nothing). Worth a separate spike if we land here.

3. **Should self-profile injection be \"diff only\" instead of full?** I.e., on follow-up, inject only profile lines that changed since the prior turn. More complex (need diff state) but maximizes savings if profile is large and stable.

4. **Do we want a per-turn budget cap?** Hard cap total injected memory_context bytes per turn at e.g. \`LARK_MEMORY_INJECT_BUDGET_BYTES\`, prioritizing profile > thread_episode > chat_episode > skill. Simple, predictable, complements the per-section caps.

5. **Interaction with `<memory_context>` security envelope.** Reducing injection must not weaken the envelope's untrusted-data marker — every block we *do* inject still needs the envelope, no exceptions on the \"fast path\".

## Non-goals

- Per-thread session isolation (separate architectural change, would require a much bigger refactor — out of scope for this issue).
- Replacing memory_context with retrieval-augmented summarization (different design discussion).
- Modifying the buffer / distiller flush behavior.

## References

- src/channel.ts:1475 (\`enrichWithMemory\`)
- src/config.ts:162-175 (current memory caps)
- src/memory/file.ts (\`searchEpisodes\`, \`searchSkills\`)

IS908:
Closes #189.

Implements the redesign agreed in the [#189 discussion](https://github.com/IS908/claude-lark-plugin/issues/189#issuecomment-4691146806): a single **content-hash dedup** mechanism replaces the original per-block-type follow-up rules table.

## What changed

`enrichWithMemory` used to inject the same-shaped `<memory_context>` slab (~7–10 KB: profile, mentioned profiles, episodes, skills) on **every** inbound message. On a hot thread the prior turn's copy is still in the main loop's conversation history, so re-injection was pure duplication accumulating in history until compaction.

Now each candidate block carries a stable `dedupKey` and is injected **iff**:
- the key was never seen in the `(chatId, threadId)` scope, OR
- its content hash changed, OR
- its last injection is older than `LARK_MEMORY_DEDUP_WINDOW_MS` (default **30 min**, `0` disables → exact pre-1.3.0 behavior).

## Design decisions (mapped to the issue discussion)

| Discussion point | Implementation |
|---|---|
| #1 hash dedup replaces rules table | `EnrichmentDedup.filter()` in [src/enrichment-dedup.ts](../blob/feat/189-enrichment-dedup/src/enrichment-dedup.ts) — one mechanism, no per-type rules |
| #2 window ≫ 5 min + stubs | Default 30 min (rationale in config comment: precondition is "copy still in history", not cache TTL). Suppressed `profile`/`mentioned_profile` render a ~150-byte envelope-wrapped `unchanged` stub so omission ≠ "no profile exists"; episodes/skills omit silently |
| #3 group multi-user hole | `dedupKey` embeds profile owner id — B's first message in a hot thread injects B's profile while A's stays suppressed (smoke case 7) |
| #4 setCaller ordering trap | Avoided entirely: dedup keeps its own state, no IdentitySession coupling |
| #5 skill intent classifier → dedup | Same-slug skill suppressed within window via `skill:<slug>` key; intent gating deferred |
| #7 measurement hook | Per-turn `[enrich-dedup]` line in debug.log: injected/suppressed counts + bytes + stubs |
| open Q3 (profile diff) | Subsumed: hash unchanged → skip; changed → full re-inject |
| open Q5 (envelope security) | Every emitted part **including stubs** goes through `wrapEnrichmentSection` — no #114 exceptions on the suppressed path (smoke cases 11/13) |

**Absolute TTL, not sliding**: suppression never refreshes the injection timestamp, so a continuously-hot thread still re-grounds every block once per window. This bounds staleness against Claude-Code-side compaction / `/clear` / restart, which the plugin cannot observe (per the [#190 feasibility verification](https://github.com/IS908/claude-lark-plugin/issues/190#issuecomment-4691202188)). Smoke case 5 locks this in.

State is in-memory only (LRU: 2000 scopes × 64 keys); plugin restart ⇒ full re-injection — the desired recovery semantics for issue open Q1.

## Scope notes

- Doc-comment and cronjob paths are unaffected (they don't run `enrichWithMemory`).
- Deferred as orthogonal until the measurement hook shows need: thread-prioritized chat-episode trimming (`LARK_MEMORY_THREAD_PRIORITIZE`), per-turn byte budget (open Q4).
- Default behavior changes (dedup ON at 30 min). Opt-out: `LARK_MEMORY_DEDUP_WINDOW_MS=0` — flagged in CHANGELOG migration notes.

## Verification

- `npm test` — full suite green, including the new `scripts/enrichment-dedup-smoke.ts` (14 cases: disable switch, hash-change re-inject, window expiry, absolute-TTL semantics, scope isolation, group multi-user keying, LRU eviction, ordering, stub-through-envelope incl. hostile label, silent episode omission).
- `bash scripts/check-doc-surfaces.sh` — `LARK_MEMORY_DEDUP_WINDOW_MS` documented in both READMEs; remaining flags are pre-existing drift on `main` (`LARK_NAME_CACHE_*`, `LARK_PROFILE_DISTILL_*`).


IS908:
## Review loop summary (3 rounds, per pre-merge process)

**Round 1 — self-review after dry-run** (`npm run start -- --dry-run` ✅): found 2 issues, fixed:
- `filter()` recorded injections before forwarding, with no recovery if the forward failed → added `invalidateScope` + try/catch in the enqueue callback.
- Hash truncation widened 64 → 128 bit.

**Round 2 — two independent review agents (correctness/concurrency + security)**, 1 major + 7 minor/nit, all fixed:
1. **[major]** Round-1's catch was dead code for transport errors — the index.ts handler swallows notification failures and returns normally. Fixed via the `MessageHandler → Promise<void | boolean>` contract: handler returns `false` on swallowed failure, IM path invalidates; the doc-comment / reaction / flush call sites ignore the return value and are untouched.
2. Inner-cap eviction would evict the always-suppressed profile entry first → suppressed hits now refresh insertion position (absolute TTL untouched, smoke 17).
3. Duplicate `@X @X` produced a same-turn self-referential stub → mentions deduped by open_id.
4. Vacuous stub-label assertion → exact `label="self:ou_a · unchanged"` check.
5. Inner cap / boundary untested → cap made injectable, smoke 17/18 added.
6. Stats overstated savings → `RenderStats.stubBytes`, debug line reports stub overhead.
7. Scope-key `#` shape could theoretically alias differently than the queue's `::` partition → aligned to MessageQueue's exact key shape.
8. Window unbounded (`9e15` ⇒ never re-ground) → clamped to 24 h; `<= 0` now logs a stderr breadcrumb.

Security review verdict: no critical/major — suppression provably limited to byte-identical re-injection, keys bound to server-resolved identities, every degradation path (eviction/invalidation/config error) fails toward re-injection, never false suppression. Two **pre-existing** `prompts.ts` hardening items (whitespace-variant close tags in `escapeEnvelopeBody`, label newline stripping) exist on `main` and are deliberately out of scope — tracked separately.

**Round 3 — fix-delta verification (fresh agent)**: verdict functionally CLEAN; 2 stale comments corrected. Hand-traced the Map insertion orders for smoke 17, exhaustively enumerated filter→invalidate paths for the delivery contract, confirmed boundary semantics are pinned by smoke 18.

Final state: `npm test` 113/113 ✅ (smoke now 18 cases), typecheck ✅, dry-run ✅, doc-surface drift check shows only pre-existing items.

PR #192, opened 2026-06-12
feat(session): session-health nudge — semi-automatic /compact at idle boundaries (v1.3.2, closes #190)

### LINKED ISSUE #190: memory: idle-triggered /compact + fallback /clear with Agent/Todo state persistence to bound accumulated history cost

IS908:
Closes #190.

Implements the re-scoped design from the [#190 feasibility verification](https://github.com/IS908/claude-lark-plugin/issues/190#issuecomment-4691202188) and [shipping plan](https://github.com/IS908/claude-lark-plugin/issues/190#issuecomment-4692340258): the issue's idle-boundary state machine survives intact — only the actuator changes from "inject a slash command" (verified impossible: hooks observe only, Skill excludes built-ins, no SDK compaction API, parallel resume unsafe) to **the operator, prompted at exactly the right moment**.

## What ships

**1. Measurement — Stop hook** ([hooks/enforce-lark-reply.mjs](../blob/feat/190-session-nudge/hooks/enforce-lark-reply.mjs))
`writeSessionStats` runs right after transcript load (so every early-exit path records) and stores the session's **exact** current context size — last main-loop assistant `usage` (`input + cache_read + cache_creation`), sidechain entries skipped — into `session-stats.json`, keyed per `session_id` (the bot megasession and dev sessions share this hook). 48 h / 32-entry pruning, atomic tmp+rename, own try/catch: **stats can never influence the hook's verdict** (case S5 locks this in). Ground truth from the reference deployment: ~0.9 M tokens measured this way.

**2. Nudge — `SessionHealthMonitor`** ([src/session-health.ts](../blob/feat/190-session-nudge/src/session-health.ts))
60 s unref'd tick; gating order `enabled → stats → close-on-drop → threshold → rearm-floor → exhausted → ladder-due → idle → quiet → retry`, each observable as a `TickOutcome` for tests. Fires an owner DM with a `[n/4]` rung counter: token count, idle duration, suggestion to type `/compact` (or `/clear`) **now, at the idle boundary** — instead of fullness-driven auto-compaction firing mid-burst.

**Exponential reminder ladder** (review follow-up; replaces the original flat 6 h cooldown):
- `LARK_SESSION_NUDGE_COOLDOWN_MS` is the **base** (default **2 h**): after the n-th unanswered nudge the next is due `base × 2^(n-1)` later → 0 / +2 h / +6 h / +14 h cumulative undelayed. Rungs anchor on the **actual** previous send, so a busy-channel delay shifts the schedule — no catch-up double-fire.
- Hard cap **4 nudges per episode**; with the 2 h base, +14 h is the last rung that can still see fresh stats inside the 24 h stats window (further doubling would be dead code).
- **Episode close**: next Stop-event measurement ≥30% below the last-nudged value (or below threshold) ⇒ operator compacted. A **1.25× re-arm floor** prevents the instant re-nudge when the post-compact level still sits above threshold.
- **Re-arm**: exhausted episode re-arms on ≥25% NEW accumulation; 24 h of ladder silence auto-resets stale state.
- DM failure → 15 min retry backoff that never consumes a rung.

Other mechanics: idle = no forwarded message (IM / doc-comment / reaction / cronjob injection) for `LARK_SESSION_NUDGE_IDLE_MS`, via `noteInbound()` piggybacking on the existing index.ts handler; quiet = `MessageQueue.pending === 0`; startup guard waits a full idle window; the nudge DM is intentionally NOT in `BotMessageTracker`.

**3. Prevention — `mcpServerInstructions`**: the host session is told to delegate heavy multi-step channel tasks to subagents, keeping tool transcripts out of main-loop history (the dominant share of the 0.9 M).

## Knobs (all opt-in, default off)

`LARK_SESSION_NUDGE_ENABLED` (requires `LARK_OWNER_OPEN_ID`), `LARK_SESSION_NUDGE_TOKEN_THRESHOLD` (400 k), `LARK_SESSION_NUDGE_IDLE_MS` (30 min), `LARK_SESSION_NUDGE_COOLDOWN_MS` (ladder base, 2 h), `LARK_SESSION_STATS_PATH`.

## Verification

- `npm test` fully green: [session-nudge-smoke.ts](../blob/feat/190-session-nudge/scripts/session-nudge-smoke.ts) (18 cases: gating order, startup guard, inbound reset, full 0/+2h/+6h/+14h schedule with rung counters + exhaustion, delayed-rung shift, drop-close + re-arm floor, below-threshold close, regrowth re-arm, 24 h auto-reset, stale/future/garbage stats robustness, send-failure backoff, start/stop idempotence) + hook suite 124 cases incl. S1–S6 (exact usage sum, no-usage no-file, corruption recovery, sidechain skip, **verdict isolation under unwritable stats path**, prune semantics). The whole hook suite defaults `LARK_SESSION_STATS_PATH` into the test tmp dir so it never touches the real sidecar.
- Dry-run + MCP stdout-cleanliness + doc-surface drift check: pass; all 5 new env vars documented in both READMEs.
- Version triplet + lockfile at 1.3.2 (per the v1.3.1 release checklist), CHANGELOG entry included.

## Explicitly out of scope

- Autonomous `/clear` / process restart — permanently dropped (session shared with operator's terminal work).
- Upstream feature request for a programmatic compact trigger — drafted separately; #190's snapshot/restore design stays parked until that exists.


IS908:
## Review loop summary (2 rounds, converged)

**Round 1 — dry-run ✅ + two independent review agents** (correctness/concurrency + security): 13 findings, all fixed in 53544b5:
- **[major ×3]** 24h time-reset scoped to the exhausted state only (was wiping live re-arm floors → zero-regrowth re-nudges, and mid-ladder state → 4-per-day instead of 4-per-episode); per-session episode identity (`episodeSessionId`) so a heaviest-entry flip can't fire spurious cross-session closes; prompt-cronjob injections now feed the idle gate via `SchedulerOptions.onActivity` (they bypass the message handler).
- Plus: single-flight tick guard, `optionalPositiveNumber` + 60 s base floor, `isQuiet` extended with IM ack reactions and de-overstated, nudge DM through `sanitizeOutboundText` (security finding — repo convention), hook hardening (zeroed-usage fallback scan, future-ts prune, orphan-tmp sweep, `.json`-only path override, loop-break staleness documented).

**Round 2 — fresh verification agent on the fix delta**: all 13 fixes confirmed correct (traced pre-fix reverts for every new test); 4 residual findings, resolved in 62e07c3:
- **[LOW, fixed]** re-arm floor now ratchets down after a second deeper compact (smoke 24 pins 1.2M→800k→500k→700k re-arming at 625k, not 1M).
- **[LOW, accepted + documented]** single-slot episode state: alternating-heaviest concurrent sessions can launder exhaustion into extra reminders — bounded by the 15-min attempt backoff + idle/quiet gates; a per-session map isn't worth it for a one-megasession deployment. Noted in code and CHANGELOG.
- Coverage gaps closed (orphan-sweep test S10 with live-writer protection) and doc stragglers fixed.

**Final state**: session-nudge smoke 24/24, hook suite 135/135, full `npm test` green (incl. dry-run + MCP stdout cleanliness), doc-surface drift check clean for all new env vars, version triplet + lockfile at 1.3.2 matching the CHANGELOG heading.

